PR #16, opened 2025-02-07
Update augmented steam to make the build smaller

tddebart:
I removed the images and localization folder from the build files so it is now 2.08mb instead of 6.4mb


shdwmtr:
You could also automate updating the commit sha instead of doing it manually.

PR #28, opened 2025-04-16
Adding steam-librarian plugin

luthor112:
Should be fixed now


shdwmtr:
Your `open_millennium_settings` function in python is unneeded. You can simply do. `window.open('steam://millennium', '_blank');` in JavaScript. 

Assuming this builds though, I'll merge it an you can fix that later.

PR #42, opened 2025-05-08
Adding steam-collections-plus v1.0.0

luthor112:


shdwmtr:
Just so you know, you should always make an effort to use React instead of manually creating elements. Millennium's API gives you ways to hook Steams internal components to render inside of them instead of finding it by class, and replacing it manually. 

For example, this is how millennium patches the Steam context menu.

```ts
function RenderSettingsModal(_: any, retObj: any) {
	const index = retObj.props.menuItems.findIndex((prop: any) => prop.name === '#Menu_Settings');

	const [isMillenniumOpen, setIsMillenniumOpen] = useState(false);

	if (index !== -1) {
		retObj.props.menuItems.splice(index + 1, 0, {
			name: 'Millennium Settings',
			onClick: () => {
				setIsMillenniumOpen(true);
			},
			visible: true,
		});
	}

	return (
		<>
			{isMillenniumOpen && <RenderMillennium setIsMillenniumOpen={setIsMillenniumOpen} />}
			{retObj.type(retObj.props)}
		</>
	);
}

const PatchRootMenu = () => {
	const steamRootMenu = findInReactTree(getReactRoot(document.getElementById('root') as any), (m) => {
		return m?.pendingProps?.title === 'Steam' && m?.pendingProps?.menuContent;
	});

	afterPatch(steamRootMenu.pendingProps.menuContent, 'type', RenderSettingsModal.bind(this));
};
```

This solution wont work for every component though, you'll have to manually sort through the components and reverse them. 
You can follow up on discord and I can help you if you need it.

PR #71, opened 2025-09-14
chore: update achievement groups plugin with small fix

BossSloth:
I no longer seem to have push rights to this repo, don't know if this was intentional or not. 
So I'll just open this pr for a small fix for the achievement groups plugin.


shdwmtr:
It was not intentional; must have broke it when changing some of the repo settings

PR #72, opened 2025-09-14
Add CSWatch plugin

Leon135:
Everything about this plugin is in readme.md

To put it simply, it adds a button with text cswat.ch which redirects to cswat.ch website.
This website is useful when looking up information about counter-strike players

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -52,3 +52,6 @@
 [submodule "plugins/cswatch-plugin"]
 	path = plugins/cswatch-plugin
 	url = https://github.com/Leon135/cswatch-plugin.git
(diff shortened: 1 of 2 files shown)


shdwmtr (changes requested):
Why is the plugin named "CSWatCH" instead of "CSWatch" as per their official website? 

Also, your `backend/main.py` has logs for "Leetify". Remove or rename them to prevent confusing end users.

Leon135:
Okay, everything should be now fixed and ready to go

Leon135:
Interesting error, I don't think rsync is the cause of it. I will look at it later.

Leon135:
fixed

PR #78, opened 2025-09-28
feat: add valve-internal-enabler plugin

veygax:
this plugin enables the "Valve Internal" menu within Steam

image


shdwmtr:
```js
    modules.get("34792").rV.m_Settings.bIsValveEmail = true;
    modules.get("72476").TS.DEV_MODE = true;
```

These references would break between updates. Sift through the modules manually and check for patterns to find the right module.

veygax:
huh, not my extension

veygax:
i can confirm its fully functional

PR #90, opened 2025-10-18
Rss feed in whats new

diemonic1:
A plugin for adding RSS news to whats new block


diemonic1:
add highlights in titles

Typically, the most important elements in headlines are anything in quotation marks or numbers, so they can be highlighted in a different color. If the news is not in English, Latin characters (game names, etc.) can also be highlighted.

PR #91, opened 2025-10-19
Add steam-global-launch-options as submodule

BlafKing:
Hi there! This is my first submission.
I've made a plugin that allows the user to configure any launch options that get automatically applied globally across all your Steam games.

Currently this plugin:

1. Gets the current original launch options for the game (if any)
2. Parses and merges them with your global options
3. Temporarily applies the combined options during launch
4. Restores original options after launch

Please let me know if anything requires adjustments/changing and I'll be more than happy to look into it!


BlafKing:
> Hello! Thank you for the submission!
> 
> `Millennium.callServerMethod` is deprecated, use `callable` instead. You can find docs for it in the typedoc comment above the `Millennium.callServerMethod` function.

Thanks for the heads up!
I've made another commit where I've replaced the usage of `callServerMethod` with `callable` along with some minor logging changes! :)

BlafKing:
Good point indeed! Hadn't considered that. The plugin already creates a default config file if one doesn't exist, so I've simply removed the `config.json` from the repository to prevent any overwrites during updates.

Regarding the Python to Lua transition, I appreciate the heads up!
I'll definitely look into preparing a Lua version of the backend so it's ready to switch over when the public version fully transitions from Python to Lua.

As for adding it now, that's totally up to you! No worries at all if you'd prefer to wait until after the transition is complete. I understand it's probably easier to manage things that way during this transitional period :)

PR #96, opened 2025-11-26
feat: add Dota Stats plugin

BambooFury:
Add dotastats plugin as a submodule.

Submodule: plugins/dotastats
URL: https://github.com/BambooFury/Dota-Stats.git


shdwmtr:
We won't be adding any new plugins until Lua support is released. As described in the discord and the plugin template, we are moving from Python backends to Lua. Unfortunately, there is no benefit of accepting a Python plugin right now if it will just have to be ported to Lua right after.

shdwmtr:
I don't understand

jcdoll:
Hi BambooFury -
Fellow plugin author and Dota enjoyer here :).

The Lua transition is done from what I can tell. I just released a plugin in Lua a few days ago. After initially writing the backend in python I saw this thread and ported my backend to Lua. It's not too bad.

You can use my plugin as a reference in case useful:
https://github.com/jcdoll/hltb-millennium-plugin

Also of course the plugin template, which is also in Lua:
https://github.com/SteamClientHomebrew/PluginTemplate

BambooFury:
I remade the plugin in Lua.

shdwmtr:
Your PR adds your repo at [1f263cf](https://github.com/BambooFury/Dota-Stats/tree/1f263cfe9c893ab431403d6b7de4cca7dc6cf628) which isn't a valid commit.

BambooFury:
I'll decide everything now

BambooFury:
I added webkit, it was just in .gitignore

shdwmtr:
Your setup is still wrong. Check the plugin template, use that as a reference.

BambooFury:
I completely redesigned the plugin structure. If anything is wrong, please describe the problem in detail so we can solve it.

diemonic1:
I checked and the plugin works fine for me

PR #100, opened 2025-12-31
Add Change Window Params plugin

ricewind012:


shdwmtr:
Does this build with the latest ttc or does it need your patches. If so, PR your patches.

PR #106, opened 2026-01-08
Add Apps-buttons

diemonic1:
This plugin allows you to add additional buttons to the Steam client, which allow you to launch any programs on your computer or open websites Millennium buttons open two links on the Internet - https://steambrew.app/plugins and https://steambrew.app/themes At any time, at the request of the user, they can be changed or deleted through the settings.json file They are inserted solely as an example of what this plugin can be used for - if the user does not need these buttons, he can remove them.

The program allows you to launch not only third-party websites through buttons, but also local programs. For this, an analog of the os.execute command is used ("cmd /c launch "" /B "startup program""), which allows you to launch the application without opening the console window.


shdwmtr:
I feel like I already responded to this PR, but maybe I forgot to click send. 

Don't use cmd commands, its *highly* insecure and *not* cross platform. In javascript, Steam provides an API to open folders and websites under `SteamClient.System`.

PR #113, opened 2026-01-19
Update HLTB for Steam to v1.1.1

jcdoll:
Updates the HLTB for Steam plugin submodule to v1.1.1.

Minor update to make it easier for users to contribute name fixes.

## Changes
- Changed name fix order to substitute -> sanitize -> substitute (supports raw Steam names with symbols)
- Fixed cache to always refetch misses so name fixes take effect immediately
- Added name fixes: Borderlands GOTY, Civilization VI, Batman Arkham Asylum GOTY, Napoleon Total War


shdwmtr:
You could keep your current algorithm, and when you think you found the game, check its HLTB id against their price API: `https://howlongtobeat.com/api/price-checks/${id}`. From there, you can see if the Steam appid matches the appid in the response, and if not retry. 

Of course there is more overhead here, but it would be a lot more accurate and likely would never require more than 2 tries.

jcdoll:
Check SteamID for validation - 100% agreed, I already do that.

https://github.com/jcdoll/hltb-millennium-plugin/blob/92eaa96e38a0b6e65b42b0d16f712ddd26cacd8d/backend/hltb_match.lua#L68

Btw IsThereAnyDeal API has some HLTB data but it is for internal use only. I've reached out to their team but it is going to remain for internal (first-party) use only for the foreseeable future.

shdwmtr:
You can actually query your entire Steam library at once too, with appid's provided.

**Endpoint:** https://howlongtobeat.com/api/steam/getSteamImportData
**Post data:** 
```json 
{ 
  "steamUserId": "your profile id, ex: ShadowMonster70",
  "steamOmitData":0
}
```

I don't know *everything* you need, but this seems like a solid solution.
This doesn't provide *all* the stats you might want, but at least it perfectly maps appid to HLTB id.

jcdoll:
Got it.

I'll play around with that new endpoint. If the current user profile is public we could just use that I believe. And then for non-public profiles, which I want to support, that endpoint could help to automate the name correction lookup stuff.

All definitely in the future work bucket, needs some experimentation.

PR #114, opened 2026-01-19
Update HLTB for Steam to v1.2.0

jcdoll:
Updates the HLTB for Steam plugin submodule to v1.2.0.

## Changes

This is a major feature update:

- **Steam Import Cache**: For users with public Steam profiles, the plugin now fetches all Steam-to-HLTB ID mappings at startup. This enables direct lookups by HLTB ID instead of relying on name-based search, significantly improving match accuracy.
- **Direct ID Fetch**: When a cached HLTB ID exists, game data is fetched directly by ID (fast path, guaranteed match)
- **Cache refetch**: Fixed cache to always refetch misses so name fixes take effect immediately
- **Name Fixes**: Now includes 250+ game name corrections
- **Test Coverage**: Added Lua unit tests for Steam import and game fetch APIs

Background discussion here: https://github.com/SteamClientHomebrew/PluginDatabase/pull/113


jcdoll:
Yeah I saw that partway through the python -> lua rewrite but chose not to use it for a few reasons:

1) unit tests run outside of millennium for CI via `busted`
2) the utils code is resuable so i can use it in scripts to test the matching performance and the scraping stuff i did today

It's simple code and so get some upsides this way

PR #159, opened 2026-04-07
Add Claude Code GitHub Workflow

shdwmtr:


shdwmtr:
@claude review this

ArjixWasTaken:
@claude review this

Edit: looks like it only runs for those with write access, good

shdwmtr:
Yeah I tried adding the bot so it would review plugins faster than me but its clearly too much work.

PR #160, opened 2026-04-10
Add CSRep Plugin

christiangrothaus:
A plugin that adds a CSRep link to every user profile.

https://github.com/christiangrothaus/cs-rep-plugin


BambooFury:
As far as I know, one of the developers is already planning to add such a plugin.
https://github.com/SteamClientHomebrew/PluginDatabase/pull/134

christiangrothaus:
> As far as I know, one of the developers is already planning to add such a plugin. #134 

Okay, are you guys against having multiple plugins that achieve the same goal? I understand if you are since you would have to verify every update. If you are I will go ahead and close this out.

PR #162, opened 2026-05-04
added steam-OwO plugin

rsa17826:
i dont know what i am supposed to put here as the added plugin has a readme talking about it, but if its suposed to just be the same thing here, then readme is here [https://github.com/rsa17826/steam-OwO/blob/main/readme.md](https://github.com/rsa17826/steam-OwO/blob/main/readme.md)


ricewind012:
wtf

PR #180, opened 2026-06-02
Add Game News

YannisFouzi:
Adds **Game News** — a Millennium plugin for the Steam Desktop client that adds a **NEWS** button to the header, opening a personal feed of news for the games you follow, full-page inside Steam (with follow management, native toasts and settings).

- Repo: https://github.com/YannisFouzi/steam-game-news-millenium
- Backend type: Lua
- Privacy policy: https://gamenews.up.railway.app/privacy

Added as a submodule under `plugins/game-news`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -73,3 +73,6 @@
 [submodule "plugins/gratitude"]
 	path = plugins/gratitude
 	url = https://github.com/BlythT/Gratitude-Millennium-Plugin
+[submodule "plugins/game-news"]
+	path = plugins/game-news
+	url = https://github.com/YannisFouzi/steam-game-news-millenium


YannisFouzi:
Updated the submodule to **v1.2.0** — adds a follow bell on a game's store page (follow/unfollow that game's news in one click, next to the wishlist button), via the plugin's webkit context.

YannisFouzi:
Updated the submodule to **v1.2.1** — the store-page follow bell now sits next to Steam's *Follow* button (instead of the wishlist button), so it also appears on games you already own.

YannisFouzi:
Updated the submodule to **v1.2.2** — fixes the store-page follow bell, which never appeared in v1.2.1 because it targeted a non-existent `#follow_button` id. It now uses the real `.queue_control_button.queue_btn_follow` selector.

PR #188, opened 2026-06-14
Add What's New RSS Ticker

CasketPizza:
I wrote a scrolling animator for  What's New, and also couldn't get the existing RSS plugin that only allows a single source to work, so I added my own version with extras.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -73,3 +73,6 @@
 [submodule "plugins/gratitude"]
 	path = plugins/gratitude
 	url = https://github.com/BlythT/Gratitude-Millennium-Plugin
+[submodule "plugins/whats-new-rss-ticker"]
+	path = plugins/whats-new-rss-ticker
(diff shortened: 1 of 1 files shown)


ricewind012:
there is no source code though, and why is a theme required ?

madmaxgrey:
great idea for the plugin but its extremely bad, your limiting the amount of people who can even use this plugin, 
plugins should be able to work regardless if a theme is applied or not,

CasketPizza:
> there is no source code though, and why is a theme required ?

i uploaded everything i did onto its github page though?

> great idea for the plugin but its extremely bad, your limiting the amount of people who can even use this plugin, plugins should be able to work regardless if a theme is applied or not,

Its because I have no clue what I'm doing honestly. I just scrambled this together and thought it worth sharing.

Apparently it does work without a skin, just re-tested.

CasketPizza:
I have now added the frontend code source and updated the readme to correct the skin info.

PR #192, opened 2026-06-20
Add Game Backlog plugin

EullerAron:
Game Backlog — a plugin to track your Steam library status. Mark games as Completed, Playing, Plan to Play or Dropped; covers get a colored ribbon and each status maps to a native Steam collection (so you can also bulk-mark with Steam's multi-select).
Repo: https://github.com/EullerAron/game-backlog

Tested on Windows with the latest Millennium. Built frontend is committed under .millennium/.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -73,3 +73,6 @@
 [submodule "plugins/gratitude"]
 	path = plugins/gratitude
 	url = https://github.com/BlythT/Gratitude-Millennium-Plugin
+[submodule "plugins/game-backlog"]
+	path = plugins/game-backlog
+	url = https://github.com/EullerAron/game-backlog


wadforth:
No hate, but why is this needed if you have collections?

EullerAron:
Yeah, fair enough haha. The thing is it actually runs on top of collections. Each status is just a normal Steam collection and the plugin reads/writes to them, so you don't really lose anything. You can still do it all through collections like usual, multi-select and all.
What I wanted was the little colored tag right on the cover, so I can glance at the grid and instantly see what I've beaten or dropped without clicking into anything. That's basically the whole point for me, I just like having it on the art.
So if collections already work for you, you really don't need it lol. I mostly put it up here cause a couple friends asked me to share it.

ejalxndr:
Hello, and thanks for the PR! Collections seem to already accomplish what this plugin sets out to do, so I'd agree with @wadforth when questioning its usefulness. I think for what it's worth, it unfortunately isn't.

PR #193, opened 2026-06-20
Add csst-at-extension

TOR968:
# CSST.at

plugin that injects a csst.at profile button into Steam community profile pages

зображення

зображення

### Developer

### Plugin Functionality

### Backend Configuration

* **Yes**: I use a standard Millennium **lua** backend in my plugin.
* **No**: I use **custom binaries** that or rely on other FOSS projects that aren't written directly using Millennium's **lua** backend. 

### Community Contribution

---

## Testing Instructions

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -73,3 +73,6 @@
 [submodule "plugins/gratitude"]
 	path = plugins/gratitude
 	url = https://github.com/BlythT/Gratitude-Millennium-Plugin
+[submodule "plugins/csst-at-extension"]
+	path = plugins/csst-at-extension
+	url = https://github.com/TOR968/csst-at-extension


ejalxndr:
We are choosing to deny this plugin. It is overly complicated for what it is trying to accomplish, and the codebase has the same issues as the last reviewal. For a plugin like this, you'd simply need a webkit - I'm not 100% sure why you have this whole "injection" setup.

TOR968:
@shdwmtr The whole problem is that the plugins weren't working for several weeks because of WebKit. I reworked everything to use injection, but when I started creating a PR, WebKit started working again. To avoid having to delete what I'd already done, I added an option to choose between them—WebKit is set as the default, and CDP is there as a backup in case it stops working again. The system load is the same as what I saw in the Millennium Plugins menu.

PR #218, opened 2026-08-05
Add Date added

Norphirion:
# Date added

Adds an **Added** card to the stats row on a game's library page, next to Cloud status, Achievements and Size on disk, showing the date the game entered the user's library.

Steam records this date but never surfaces it anywhere in the client. It makes it easy to tell a years-old backlog entry from something picked up in the last sale.

The date is read from `rt_purchased_time` on Steam's local app overview: no network access, no account linking, no configuration. Class names are resolved at runtime via `findClassModule`, so the card keeps Steam's own styling rather than hardcoding hashed class names.

### Developer

### Plugin Functionality

### Backend Configuration

### Community Contribution

Feedback left after source audits and live Steam Client Stable testing:
[Sortium - PR #231 ](https://github.com/SteamClientHomebrew/PluginDatabase/pull/231)
[Non-Steam Store Link - PR #223](https://github.com/SteamClientHomebrew/PluginDatabase/pull/223)
[Recent Chats - PR #221](https://github.com/SteamClientHomebrew/PluginDatabase/pull/221)
[Steam Wrapped - PR #220](https://github.com/SteamClientHomebrew/PluginDatabase/pull/220)

---

## Testing Instructions

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -73,3 +73,6 @@
 [submodule "plugins/gratitude"]
 	path = plugins/gratitude
 	url = https://github.com/BlythT/Gratitude-Millennium-Plugin
+[submodule "plugins/date-added"]
+	path = plugins/date-added
+	url = https://github.com/Norphirion/date-added


Shightrox:
Disclosure: I am the author of #224 and tested this PR as part of the Community Contribution requirement.

Tested on Windows 11, Steam Client Stable, Millennium 3.4.0, using the exact pinned plugin commit `b349cac37510647db747971aa5e6267b461d1d56`.

### Runtime results

- The plugin loaded without a backend and without making its own network requests.
- On CS2 (`730`) and Dota 2 (`570`), where Steam exposed a positive `rt_purchased_time`, the injected cards had non-empty text and followed the active app route correctly.
- On two owned games where that field was absent (`105600` and `1604000`), no empty/stale card was rendered.
- I found no Date added warnings or errors in `webhelper_js.txt` during navigation.
- Steam exposed two visible `GameStatsSection` hosts on both tested pages, and the plugin inserted a card into both. This may be intentional for responsive layouts, but it is worth manually confirming that the user never sees a duplicate in any library layout.

### Issues found

1. **The PluginDatabase build workflow is not currently reproducible from a clean checkout.** Under Node 20.20.2, using the database's pnpm flow, the repository has no tracked `pnpm-lock.yaml` (only `package-lock.json`). `pnpm install` therefore resolved `@steambrew/ttc` from `^3.2.6` to `3.3.7` and its TypeScript peer to `7.0.2`; `pnpm run build` failed, including `TS5108` because `moduleResolution=node10` is removed in TS 7, followed by a `rollup-plugin-inject-process-env` resolution error. Please commit a pnpm lockfile or pin a known-compatible toolchain and verify a clean Node 20 `pnpm install && pnpm run build`.
2. **There is no unload cleanup.** `start()` creates a permanent one-second interval, but the plugin returns no `onDismount`; `removeCards()` removes React hosts without calling `root.unmount()`. Disabling/reloading the plugin can therefore leave the interval and React roots alive. Returning a cleanup function that clears the interval, unmounts each root, and removes the hosts would make hot reload/disable safe.
3. `plugin.json` reports `1.0.1`, while `package.json` reports `1.0.0`.
4. The `$schema` URL currently returns 404 (`src/sys/plugin-schema.json` has moved), so editor validation is not working.
5. The package declares MIT and the README says MIT, but the pinned repository has no `LICENSE` file. Adding the actual license text would make the licensing claim verifiable.

The core feature behaved correctly in the routes I tested; the clean-build and lifecycle points are the main things I would address before publishing.

Norphirion:
Thank you for this review @Shightrox, the depth of it is genuinely appreciated. Testing on real apps, checking the negative cases (`105600`, `1604000`), reading `webhelper_js.txt` and actually attempting a clean build is far more than a rubber stamp, and every one of your five points was valid.

Some context that may be useful: I'm not a developer. This plugin was written with Claude Code, and I reviewed and tested the result rather than authored it line by line. So I apologise for the rough edges you had to find, especially the build and lifecycle ones, which are exactly the kind of thing I lacked the background to catch. I'll do my best to address anything else you spot.

All five points are fixed in the updated commit. Details below.

## 1. Clean build under pnpm

This was the real bug, and your diagnosis was exactly right. `typescript` was never declared as a dependency at all - it was only present transitively - so `pnpm` was free to satisfy `@steambrew/ttc`'s peer with TypeScript 7, which is where `TS5108` came from.

Fixed by:

- Pinning every devDependency to an exact version (no `^`).
- Declaring `typescript` explicitly at `5.9.3`, plus a `pnpm.overrides` entry so it cannot be hoisted to a different major anywhere in the tree.
- Committing a `pnpm-lock.yaml`.
- Switching `moduleResolution` from `node` to `bundler`, which both TypeScript 5 and 7 accept — so the build no longer depends on the resolver that broke for you.
- Dropping `@steambrew/api` and `@steambrew/webkit`, which were declared but never imported.

Verified rather than assumed: from a clean checkout containing only the tracked files, `pnpm install --frozen-lockfile` followed by `pnpm run build` succeeds and produces a bundle byte-identical to the committed one. That was with pnpm 9 on Node 24 rather than your Node 20.20.2, so if it still misbehaves on 20 I'd very much like to know.

## 2. No unload cleanup

Correct, and worse than it looked: the plugin returned nothing at all from `definePlugin`, with a type cast to force that past the compiler. `Plugin` has supported `onDismount` all along.

The plugin now returns a proper `Plugin` object whose `onDismount` clears the interval, unmounts every React root and removes the hosts. Two related details:

- The roots were held in a `WeakMap`, which cannot be enumerated at teardown, so it is now a `Map`.
- That change would have introduced a slower leak of its own, since React detaches our hosts whenever it re-renders the stats section. Each tick now prunes roots whose host is no longer connected and unmounts them.

Only `icon` is returned, so no settings panel is mounted. Millennium requires `title`, `icon` and `content` together for that, and this plugin has nothing to configure.

## 3. Version mismatch

Fixed; `plugin.json` and `package.json` both read `1.0.1`.

## 4. `$schema` 404

Removed rather than repointed, and I think this one is worth flagging upstream.

`src/sys/` no longer exists in the Millennium repository at all, and I could not find a `plugin-schema.json` anywhere else in it, it appears to have gone during the move to Starlight. The official `PluginTemplate` and the other published plugins I checked still carry the same dead URL.

Since `$schema` is editor-only metadata with no runtime effect, dropping the line seemed better than pointing at a URL I'd be guessing at. Happy to add the correct one if it exists somewhere I didn't look.

## 5. Missing LICENSE

Added : the MIT text, matching what `package.json` and the README already claimed.

## On the two `GameStatsSection` hosts

Good catch, and worth documenting. I measured both on a game page:

- One sits under an ancestor with `opacity: 0` — it is the sticky `PlayBar` that fades in when you scroll. `checkVisibility()` returns `false` for it, and `elementFromPoint` at its centre does not land inside it.
- The other has nothing hiding it and is the one actually on screen.

So only one is ever visible and the user never sees a duplicate. Injecting into both is deliberate: if the card only went into the in-page section, it would disappear the moment the sticky bar takes over on scroll. I've added a comment in the source explaining this so it doesn't look accidental to the next reader.

Thanks again, the plugin is in noticeably better shape than before your review.

ejalxndr:
The `.millennium/` directory should be ignored by git. The PluginDatabase will create that automatically.

Norphirion:
Thanks for catching it, I had this backwards.

`.millennium/` is now in `.gitignore` and untracked in both of my plugins.

The README no longer tells people the bundle has to be committed.

PoorPocketsMcNewHold:
# ~~Project contain a malicious NPM package.~~
```
󰣇  …/date-added   main   v26.7.0   11:00  
 npm install
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@7.2.3: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me
npm warn deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
npm warn deprecated glob@11.1.0: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me
npm warn deprecated glob@8.1.0: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me
npm error code E403
npm error 403 403 Forbidden - GET https://registry.npmjs.org/fs/-/fs-0.0.1-security.tgz
npm error 403 In most cases, you or one of your dependencies are requesting a package version that is forbidden by your security policy, or on a server you do not have access to.
npm error A complete log of this run can be found in: /home/pm/.npm/_logs/2026-08-17T09_00_36_514Z-debug-0.log

✗ Malicious package blocked

  - fs@0.0.1-security
    Reference: https://app.safedep.io/community/malysis/01K32N96RPJWBP3M37FVWGWXWF

✗ PMG: 197 packages analyzed, 1 blocked
↳ pmg: npm exited with code 1
```

~~https://github.com/Norphirion/date-added/blob/main/pnpm-lock.yaml#L1307~~

~~https://github.com/Norphirion/date-added/issues/1~~

Norphirion:
Thanks for running a supply-chain scan @PoorPocketsMcNewHold 
I'd much rather you flagged this than stayed quiet. I've dug into it, and I believe it's a false positive, but the inconvenience on your side is real. Evidence below so you can check rather than take my word for it.

**What the package is.** `fs@0.0.1-security` is npm's own placeholder for the squatted `fs` name. Its repository is `github.com/npm/security-holder`, and its README says npm holds the name "to avoid malicious use". The published tarball contains exactly two files — `package.json` and `README.md` — with no code, no install scripts, and not even the `index.js` its own manifest points at. It also cannot be loaded at runtime: `fs` is a Node builtin, so `require('fs')` resolves to the core module no matter what sits in `node_modules`.

**Where it comes from.** Not from this plugin:
date-added@1.0.1 └── @steambrew/ttc@3.3.7 └── [fs@0.0.1-security](mailto:fs@0.0.1-security)

`@steambrew/ttc` is Millennium's own build toolchain, and it declares `fs` as a direct dependency. I checked every version published to npm — all 32 of them, from `1.0.0` through `3.3.7` — and the dependency is present in every single one. There is no version of `ttc` that avoids it.
**Why it isn't specific to my plugin.** Other plugins already in the database pull the same thing, because they use the same compiler:
- `size-on-disk` — `@steambrew/ttc@^2.8.7`
- `non-steam-playtimes` — `@steambrew/ttc@^2.8.7`
- `steam-share` — `@steambrew/ttc@^3.2.6`
- `extendium` — `@steambrew/ttc@~3.2.6`
The official `PluginTemplate` declares `@steambrew/ttc@^3.2.6` too, so anyone following it from scratch inherits it as well.
I genuinely can't tell you why it surfaced here and not on those — that's a question about which reviews got scanned with which tooling, and I have no visibility into it. What I can show is that the package is not something I added, and not something I can remove from my own manifest.
**What I think should happen.** The clean fix is upstream: `ttc` shouldn't list `fs` at all, since it's a Node builtin and the entry does nothing but drag in a placeholder that trips scanners. I'm happy to open an issue on the Millennium repository — it would clear this for every plugin author, not just me — unless a maintainer would prefer to handle it directly.
If you'd rather I neutralise it on my side in the meantime, I can add a package-manager override so the placeholder is never fetched. I've deliberately not done that yet: it would paper over an upstream packaging bug in a way every future plugin author would have to copy without knowing why. Say the word and I'll add it.

Norphirion:
The upstream fix has landed — fs was removed from @steambrew/ttc's dependencies and the version bumped to 3.3.8 [commit link](https://github.com/SteamClientHomebrew/Millennium/commit/44e01eb437b95c6e2004aed5dc3224bfb172d00c) It isn't on npm yet, so my lockfile still pins 3.3.7; I'll bump as soon as 3.3.8 is published.

Norphirion:
Pushed 1.0.2, which addresses the three points raised by @SalvadorCorreia that created the Sortium plugin #231 , in a Discord PM:

> Now about "date-added":
> It's a very simple plugin, so it was pretty easy to try it out, I liked the idea and felt it was pretty charming.
>
> However, I found something that might quality as a bug (?).
> The placement of the Date Added text was a pretty inconsistent. I have attached a couple of pictures showing this off and if you need to see it happening I can try to record it. Also, the time the text took to show up varied a lot, and the icon next to the text wasn't the same in the version that I downloaded and the one in the screenshots you sent in the server.

Thanks for taking the time, and for describing it precisely enough to be reproducible. All three were real, and all three had the same underlying flavour: the plugin was relying on timing it does not control.

**Inconsistent placement.** The stats row is a flex container, and the card was simply appended to it. Steam renders its own cards asynchronously, and the achievements card in particular waits on a network fetch, so whether our card landed before or after it came down to which finished first. I reproduced it on a live client: on one page load the card sat at position 3, ahead of `Succès 19/50`.

The card host now carries an explicit CSS `order`, which pins it last regardless of where it ends up in the DOM. Steam's own cards leave `order` at its default of `0`, so this is additive rather than a fight over positioning. Verified by measuring rendered positions rather than DOM order: the card is now last in both the in-page stats row and the sticky play bar, even when its DOM index is earlier than another plugin's card.

**Varying delay before the text appeared.** The plugin polls for the stats section, and the interval was one second, with the first pass only running after that second had elapsed. So the wait was anywhere from nearly nothing to just over a second depending on where in the cycle the navigation landed. The interval is now 400ms and the first pass runs immediately. Some variation remains and is not removable from this side, because the card can only render once Steam has the app's `rt_purchased_time` available.

**The icon changed between versions, and looked out of place.** Two separate things were happening here.

The mismatch you saw between my screenshots and your install is a theme effect. Millennium themes restyle Steam's stat icons by class, and the card was borrowing `GameStatIcon` + `PlaytimeIcon` to get the native sizing and colour. The theme I was running keys on exactly that pair, hides the SVG and substitutes a glyph from an icon font, so my screenshots showed the theme's playtime glyph rather than the plugin's own icon. On an unthemed client you get the plugin's icon instead. That was never obvious from the screenshots, and it is a fair thing to have been confused by.

The icon itself was also genuinely wrong, which is the part worth fixing. Steam ships several icon families, and the stats row uses the square one. Measured on the client, every calendar Steam exposes is `1200x1200`, while `ShoppingCart`, `HardDrive` and `Achievement` are all `36x36`. Squeezing a 1200x1200 asset into a 26px box is why it looked like it came from somewhere else, which it did. **I have swapped the calendar for the cart icon**, which sits in the same family as the surrounding icons and reads reasonably for "entered the library".

`pnpm install --frozen-lockfile && pnpm run build` from a clean export of the tracked files reproduces my local bundle byte for byte.

senkumil:
I haven't tried it, but it looks like a very promising plugin.

Norphirion:
Hey @ejalxndr !
Do I need to do anything else foy my plugin to be approved ?

PR #219, opened 2026-08-05
Add Exclude collections in collection

Norphirion:
# Exclude collections in collection

Adds collection membership as a criterion for dynamic collections. Pick a dynamic collection, then either exclude every game that already belongs to other collections, or keep only those.

Steam's dynamic collections filter on tags, features, platforms and playtime, but not on other collections — so there is no way to express "everything I'm playing, minus everything I've already finished". This fills that gap. Rules are set per dynamic collection, stack on top of that collection's existing Steam criteria rather than replacing them, and refresh on their own when a source collection changes.

### Notes for review

The plugin never touches Steam's cloud-synced filter definition. It replaces the collection's in-memory filter object with a stand-in that inherits the same prototype, forwards every own property back to the original (MobX administration symbols included), and overrides only `Matches` and `bIsEmpty`. Disabling the plugin restores every collection exactly as it was, and nothing is written to the user's Steam Cloud data.

Only dynamic collections can be targeted: a static collection has no filter of its own, and routing one through the filtering path would discard hand-curated membership. Rules are stored in `localStorage`.

### Developer

### Plugin Functionality

  - The functionality of my plugin exists in the plugin "Collection+", but I wanted to have a standalone version of this options, with a user interface easier to use than adding some commands in the category settings

### Backend Configuration

### Community Contribution

Feedback left after source audits and live Steam Client Stable testing:
[Sortium - PR #231 ](https://github.com/SteamClientHomebrew/PluginDatabase/pull/231)
[Non-Steam Store Link - PR #223](https://github.com/SteamClientHomebrew/PluginDatabase/pull/223)
[Recent Chats - PR #221](https://github.com/SteamClientHomebrew/PluginDatabase/pull/221)
[Steam Wrapped - PR #220](https://github.com/SteamClientHomebrew/PluginDatabase/pull/220)

---

## Testing Instructions

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -73,3 +73,6 @@
 [submodule "plugins/gratitude"]
 	path = plugins/gratitude
 	url = https://github.com/BlythT/Gratitude-Millennium-Plugin
+[submodule "plugins/exclude-collections-in-collection"]
+	path = plugins/exclude-collections-in-collection
+	url = https://github.com/Norphirion/exclude-collections-in-collection


Norphirion:
Heads up: this plugin was written by the same person, with the same toolchain, as `date-added` (#218 ), so it shipped with exactly the same defects. Rather than wait for a reviewer to find them a second time, I've applied the same fixes here:

- **Reproducible build.** Every devDependency pinned to an exact version, `typescript` declared explicitly at `5.9.3` with a `pnpm.overrides` entry, `pnpm-lock.yaml` committed, and `moduleResolution` switched from `node` to `bundler` so the build survives TypeScript 7. Verified from a clean checkout with `pnpm install --frozen-lockfile && pnpm run build`.
- **Unload cleanup.** The plugin now returns `onDismount`, which stops the watch timer and restores every collection's original filter object. Previously both were left running after a disable or reload. The async startup is also guarded, so an initialisation still pending when the plugin is dismounted can no longer reinstall filters afterwards.
- **`$schema` removed.** The referenced file no longer exists upstream; see the discussion in the other PR.
- **LICENSE added** (MIT), matching the README and `package.json`.
- **Unused dependencies dropped** (`@steambrew/api`, `@steambrew/webkit`).
- Also fixed a bug I found while doing this: the settings panel still displayed the plugin's former name.

For the record on what this plugin touches, since it is more invasive than a display-only plugin: it never writes to Steam's cloud-synced collection data. It substitutes the collection's in-memory filter object with a stand-in that inherits the same prototype, forwards every own property back to the original - MobX administration symbols included - and overrides only `Matches` and `bIsEmpty`. `onDismount` puts the originals back, so disabling the plugin restores every collection exactly as it was.

I'm not a developer and this was written with Claude Code, so I'd rather over-explain the mechanism than have it look like magic. Happy to answer anything or change the approach if it raises concerns.

Norphirion:
Applied the same fix about `.millennium/` as #218

Norphirion:
Pushed 1.1.0. Two changes, one user-facing and one a bug fix.

## Rules are now set inside Steam's own filter UI

Previously the rules lived only in the Millennium settings panel, away from the filters they extend. Two buckets : **Exclude collections** and **Include collections**  are now injected into Steam's filter area, in two places:

- the filter editor of an existing dynamic collection
- the pane used to create a new dynamic collection, where the choices are held until Steam assigns the collection an id, then applied to it

Pick collections from a dropdown, click a chip to remove one. A collection already used by one bucket is greyed out in the other. The settings panel is kept — it is the only place that lists every rule at once — and both paths write through the same function.

The injection is append-only: no element of Steam's own is modified, moved or restyled, so disabling the plugin leaves the panel exactly as it was. Two layout details turned out to be load-bearing and are documented in the README:

- Steam gives its buckets an explicit CSS `order`, so ours need one too or they render first.
- The filter area is a `repeat(4, 1fr)` grid with no imposed width, which makes `1fr` resolve against **max-content**. The max-content of a wrapping chip row is every chip on a single line, so each chip added widened all four columns (425px instead of 271px with seven chips). The chip container now carries a definite `width: 0` — which contributes nothing to that measurement — plus `min-width: 100%` to fill the column once the track has been sized.

## Fix: the filter wrapper could be silently discarded

The engine wrapped a collection's `m_filter` once and then trusted its own bookkeeping. Steam rebuilds that object in at least two situations — when a collection is edited and saved, and while it finishes its own start-up after the plugin has wrapped — which dropped the stand-in while the plugin still believed it was installed. The rule then stopped applying, most visibly as stale, unfiltered collections right after launch.

Each poll now re-checks that the wrapper is still in place and reinstalls it otherwise, and the watcher starts as soon as any rule exists rather than only once a wrap has succeeded, so a collection that was not ready at start-up is picked up on a later pass.

Verified after a cold restart by comparing, for every rule, the collection's actual contents against what the native filter plus the rule should produce — exact match on all of them.

## Unchanged

Nothing about what the plugin touches has changed: the cloud-synced `filterSpec` is never written, only dynamic collections can be targeted, and `onDismount` still restores every original filter object.

`pnpm install --frozen-lockfile && pnpm run build` from a clean export of the tracked files produces a bundle byte-identical to my local build.

Norphirion:
1.1.1 — a layout fix on the buckets added in 1.1.0.

**The bug.** The filter panel grew far more than the buckets needed. With seven chips wrapping over a few lines it went from 263px tall to 663px, and it kept growing with each chip added.

**The cause.** Steam's filter area is declared `grid-template-rows: 2fr 1fr`, and Steam's own buckets occupy the `2fr` row. Anything appended to that grid lands in the `1fr` row — so every pixel of height it needs costs three, because the native row is forced to twice ours regardless of what its own content asks for. The native row was sitting at 442px while its content only needed 175px.

**The fix.** Steam has the same problem with its own Language and Genre buckets, and solves it by placing them in a sibling container outside the grid. The buckets now go into a container of their own built from that same class, so they inherit its layout and breakpoints and land just after it. Measured on the same collection, the panel stays at 263px with seven chips.

Two things fall out of that: the explicit CSS `order` the buckets needed to sort correctly among Steam's own is no longer necessary, and teardown removes the container along with them. Still append-only — no element of Steam's is modified, moved or restyled.

`pnpm install --frozen-lockfile && pnpm run build` from a clean export of the tracked files still reproduces my local bundle byte for byte.

Norphirion:
Pushed 1.1.2, which addresses the three points raised by @SalvadorCorreia that created the Sortium plugin #231 , in a Discord PM:

> Firstly, since the options you were inserting were placed inside the "sorting menu" it felt weird to me that selecting them never had any impact on the list displayed on the left-side of the screen.
>
> Secondly, it felt weird to me that if you never selected any of steam's native options, the "Save Dynamic Collection" button would never light up, making it so you can't create a collection just by including and excluding games from other collections.
>
> Lastly, I noticed that you decided to display collection names using their bare text, however, this can feel pretty weird in the steam context. Have you ever noticed that Steam always displays collection names in all caps?

Thanks for these. Two are fixed, one I am deliberately leaving alone, and answering the second one turned up a real bug.

**Collection names are now upper-cased**, in the dropdown and on the chips. You were right: `SectionName` and `InCollection` both carry `text-transform: uppercase` in Steam's own stylesheet, and our bucket labels were already inheriting it. Done through CSS rather than in the string, so the name stays intact for screen readers and copy-paste, and a collection someone happened to type in capitals no longer stands out.

**A bug your second point uncovered.** Rules stored a single mode with one list of sources, so choosing a collection in one bucket silently replaced whatever the other held. The greying-out of collections "already used by the other bucket" could never fire either, since the other bucket was always empty by construction. Rules now hold both lists at once: inclusions narrow the pool, then exclusions are removed from it, and an empty include list means the rule does not narrow anything so exclusions still work on their own. Existing rules are migrated on read. Verified against a real library: identical results before and after the migration.

**The Save button.** This one stays as it is, and I want to explain why rather than leave it looking unfinished. The button's disabled state is React state that Steam derives from its own filter definition being empty. The only ways to light it up are to write into that definition, or to fight Steam's React state from outside. The first is the problem: Steam's filter definition is cloud-synced, it has no notion of collection membership, so we would not be storing our rule there, we would be storing an unrelated decoy option purely to make the filter look non-empty. That option would then appear in Steam's own UI as a filter the user never chose, sync to every machine on the account, and survive uninstalling the plugin. Not worth it for a UI gate. Pick any one native option and the rule saves alongside it.

**The preview list.** I could not reproduce it in the collection editor, where the list and the count do follow our rules. You are right about the creation popup though: there, the preview is computed from a draft filter object that only knows about Steam's own criteria. I looked into hooking it, and it is reachable, but only by walking React's internal fiber tree, and making the preview actually refresh would mean touching the draft filter definition, which is the same object that gets saved. Both add fragility to the most invasive part of the plugin for a preview convenience, so I am leaving it. The rule applies correctly as soon as the collection exists.

Worth noting for completeness: a dynamic collection can also contain manually added games, which Steam keeps regardless of any filter. Those are not something this plugin can remove, and it should not.

`pnpm install --frozen-lockfile && pnpm run build` from a clean export of the tracked files reproduces my local bundle byte for byte.

PR #228, opened 2026-08-17
Update Easy Restart / Reload for Steam

SaiyajinK:
### Changes : 

- Fixed the **Reload Interface** action after the PluginDatabase integration by replacing `documentRef.defaultView?.location.reload()` with `window.location.reload()`.
- Fixed the plugin name display in the PluginDatabase.
- Updated the README with Millennium installation instructions.


SaiyajinK:
I think the workflow fails during the repository checkout because the `plugins/dotastats` submodule points to `https://github.com/BambooFury/Dota-Stats.git`, which currently returns a “repository not found” error.

image

PR #229, opened 2026-08-17
Update Steam Share

wopln:
Updates Steam Share to v1.1.0.

This version adds Favorites, Recent badges, drag-and-drop favorite management, and direct friend-card selection. It also keeps the history cleanup fix from the original review.

The update now pins TypeScript 5.9.3 and includes the pnpm build approvals needed by the current PluginDatabase runner. This fixes the build failure that caused Millennium to report "Invalid Build."

Release: https://github.com/wopln/Steam-Share/releases/tag/v1.1.0

Checked locally with:
- Bun TypeScript check
- Bun production build
- Clean pnpm install
- pnpm TypeScript check
- pnpm production build


ejalxndr:
Don't use `globalThis.SP_REACT`, you can import from react directly.

wopln:
Fixed in wopln/Steam-Share@fe22a68. React hooks and createElement are now imported directly from react, and the submodule pointer has been updated. The local production build and TypeScript checks pass.

PR #243, opened 2026-08-31
Add Steam Currency Converter

Jidos86:
# Steam Currency Converter

Shows an approximate price **in the currency you pick** next to any Steam price — store, cart, checkout, community market, and the in-game overlay. It detects the account's wallet currency and appends `≈ N <cur>`, computed from public exchange-rate JSON ([fawazahmed0/exchange-api](https://github.com/fawazahmed0/exchange-api), 3 mirrors, cached 6 h). Not an official regional price — a raw exchange-rate estimate for quick comparison.

Steam Currency Converter on the Steam store

Maintained fork of the existing (unlisted) https://github.com/KuroKim/steam-currency-to-rub. Differences:

- **Target currency is user-selectable** in the plugin settings (was hard-coded to RUB); all 40 Steam wallet currencies, as source *and* target.
- **Injection**: moved from a Lua `add_browser_js` `<script src>` (silently blocked by the store CSP in some regions, e.g. the Steam China store) to a `webkit/` module that Millennium loads with `Page.setBypassCSP`.
- Format-agnostic price parser (`1,234.56`, `1.234,56`, `1 199`, zero-decimal currencies).
- Full-precision rate (the original rounded to 2 decimals, breaking every sub-1 rate — USD/EUR/GBP/CNY, error up to tens of percent).
- Works on the new React cart/checkout.
- **DOM-only rendering** — `createElement` / `textContent`, no `innerHTML`, no `eval`, no remote code. Network is limited to the exchange-rate endpoints in `webkit/index.tsx`.

Settings persistence goes through the lua backend (`get_settings` / `set_target_currency` via `millennium.config`) because `usePluginConfig` / `pluginConfig` were unreliable in the current Millennium build.

No RUB/currency-conversion plugin is currently on the store.

### Developer

### Plugin Functionality

### Backend Configuration

### Community Contribution

---

## Testing Instructions

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -91,3 +91,6 @@
 [submodule "plugins/Easy-Restart-Reload-for-Steam"]
 	path = plugins/Easy-Restart-Reload-for-Steam
 	url = https://github.com/SaiyajinK/Easy-Restart-Reload-for-Steam
+[submodule "plugins/steam-currency-to-rub"]
+	path = plugins/steam-currency-to-rub
+	url = https://github.com/Jidos86/steam-currency-to-rub
--- a/.gitmodules
+++ b/.gitmodules
@@ -91,6 +91,6 @@
 [submodule "plugins/Easy-Restart-Reload-for-Steam"]
 	path = plugins/Easy-Restart-Reload-for-Steam
 	url = https://github.com/SaiyajinK/Easy-Restart-Reload-for-Steam
-[submodule "plugins/steam-currency-to-rub"]
-	path = plugins/steam-currency-to-rub
-	url = https://github.com/Jidos86/steam-currency-to-rub
+[submodule "plugins/steam-currency-converter"]
+	path = plugins/steam-currency-converter
+	url = https://github.com/Jidos86/steam-currency-converter


Jidos86:
Converting this to a **draft** — please hold off reviewing for now.

Two things are landing before it's ready:
1. Renamed the plugin `steam-currency-to-rub` → `steam-currency-converter` (id `steam_currency_converter`). The submodule pointer/URL in this PR is already updated.
2. Adding a settings panel so the target currency is user-selectable (currently hard-coded to RUB). This means moving to the `@steambrew` build (`frontend/index.tsx` + `webkit/index.tsx` + lua backend).

I'll mark it ready for review and advance the submodule pointer once that's done and tested on Stable + Beta.

Jidos86:
Out of draft and ready for review.

- Renamed `steam-currency-to-rub` → `steam-currency-converter`; submodule pointer advanced to `9aa0b08` (v2.0.0).
- Target currency is now user-selectable (settings panel); default RUB.
- Tested on Steam **Stable** and **Beta** (incl. a CN-region account).

Still open: feedback on two other plugin PRs — will follow up.

kuusei:
Disclosure: I am the author of #241 and tested this PR as part of the Community Contribution requirement.

### Fallback currency detection will mis-label JPY and SEK
Primary detection (wallet id, then meta[itemprop="priceCurrency"]) looks correct. The formatter fallback is not:

```
const SIGN_TO_CURRENCY: Record<string, string> = {
  '¥': 'CNY',
  '$': 'USD',
  // …
};
```
¥ is also JPY. kr is both NOK (id 9) and SEK (id 33); it is not in SIGN_TO_CURRENCY, so STEAM_CURRENCIES.find((c) => c.symbol === sign) returns NOK first. A Japanese or Swedish account that misses wallet + schema.org detection gets converted as CNY or NOK.

Wallet id is the right source of truth — if it is missing, skipping conversion is better than a silently wrong rate.

Jidos86:
Thanks for testing, @kuusei — good catch.

Fixed in v2.0.1 (pinned commit updated): the symbol-based formatter fallback is gone entirely. Source currency now comes only from the wallet id and the `priceCurrency` meta; if neither is available the plugin skips conversion instead of guessing. `SIGN_TO_CURRENCY` stays only as a "looks like a price" hint for the DOM scan, where an exact match isn't needed.

Jidos86:
Community Contribution — review 1 of 2: https://github.com/SteamClientHomebrew/PluginDatabase/pull/241#issuecomment-5479336981 (Sort As Display — found 2 bugs worth fixing before merge: hooks not removed on unmount, and the SetCustomSortAs hook not retrying on a failed install).

Jidos86:
Community Contribution — review 2 of 2: https://github.com/SteamClientHomebrew/PluginDatabase/pull/231#issuecomment-5479513798 (Sortium — flagged the Augmented Steam / Steam Hunters API fan-out without a User-Agent or coordination, stringly-typed rate-limit detection, a NaN-comparator sort for missing data, and an O(n²) non-atomic cache write).

Both required reviews done: #241 and #231.

kuusei:
> Thanks for testing, @kuusei — good catch.
> 
> Fixed in v2.0.1 (pinned commit updated): the symbol-based formatter fallback is gone entirely. Source currency now comes only from the wallet id and the `priceCurrency` meta; if neither is available the plugin skips conversion instead of guessing. `SIGN_TO_CURRENCY` stays only as a "looks like a price" hint for the DOM scan, where an exact match isn't needed.

Confirmed on 678e071. Formatter fallback is gone, and it skips conversion when wallet id / priceCurrency are both missing. LGTM.

Jidos86:
Pinned commit advanced to v2.0.2 (`ad4e309`). Since the last update, from user testing on a CN-region account:

- discount blocks: parse each price token separately and take the final one (was concatenating `-15% ¥98.00 ¥83.30` into a billions figure)
- render by appending a hint span, never rewriting the price element — keeps Steam's own discount badge / strikethrough / layout intact
- convert on `/search`, the discovery feed, the instant-search dropdown and the (React) inventory market price; source currency is cached once wallet-id / priceCurrency detection succeeds so pages exposing neither still work
- never annotate a struck-through original price (checked via computed `text-decoration`)
- pick up prices that load asynchronously into existing nodes

CI green.

Jidos86:
Pinned commit advanced to `0d4c29b`. Added a generic leaf-price scan: the new React store widgets (recommendation lists, feeds, capsules, the instant-search dropdown) render the price as a leaf element whose text is exactly a wallet-currency price, so those are converted directly, skipping struck originals via the computed `text-decoration`. CI green.

Jidos86:
v2.0.3 (`870c6ff`): currency detection now retries — the store home / feeds populate `g_rgWalletInfo` asynchronously, so on a fresh install prices only showed up after visiting a game page. Also the one-liner installer now bootstraps Millennium itself when it's absent (mirrors the official installer's steps). CI green.

Jidos86:
Pinned commit advanced to v2.0.4 (`a464f97`). Since v2.0.3, mostly hardening from a code review:

- installer scripts unlink junctions/symlinks instead of recursing into them (a dev symlink can't be wiped by re-running install/uninstall); Millennium config is written UTF-8 without a BOM and keeps `enabledPlugins` a JSON array
- the price sanity cap is per-currency now (`1e11` for zero-decimal currencies vs `1e8`) so it stops dropping valid high-denomination market prices
- the source-currency cache is keyed by `g_steamID` when it's exposed
- the webkit `MutationObserver` and its follow-up timer are disconnected on `pagehide`
- settings dropdown rolls back if the save call fails; CI split so `contents: write` is only on the tag/release job

CI green (build + release).

PR #248, opened 2026-09-04
Added "Autoplay Mute" and "Steam Hoverless"

SnowFallinSummer:
Adds two Steam plugins to the PluginDatabase:

- Steam Autoplay Mute — disables automatic video playback in Steam.

- Steam Hoverless — disables game hover-preview popups in Steam Library, Home, and Store.

Both plugins are maintained in their respective repositories.


ricewind012:
hoverless is just a single css rule, and the rest you have PRs opened for already

